PR #10, opened 2021-04-05
testbuild-gapps

loicometa:


nikgapps (changes requested):
Hi, you're supposed to add a new file, don't delete existing config files. In this case, rageshgapps.config should not be deleted. I would suggest you to make the changes and send a pull request again

loicometa:
> Hi, you're supposed to add a new file, don't delete existing config files. In this case, rageshgapps.config should not be deleted. I would suggest you to make the changes and send a pull request again

Ohh, alright sir. I didn't know. Will do it again sir. Thank you.

PR #19, opened 2021-04-05
Update TASEER.config

btaseer123:


11/TASEER.config
```diff
@@ -113,15 +113,15 @@ PlayGames=1
...
+GoogleFeedback=0
 GooglePartnerSetup=1
 GoogleSounds=1
 AndroidDevicePolicy=1
 GoogleKeep=0
 Books=0
 Assistant=0
 YouTubeMusic=0
-GoogleRecorder=1
+GoogleRecorder=2
```
nikhilmenghani, line 124: You've set GoogleRecorder=2, this will be skipped as the ones will value 1 are included.

PR #44, opened 2021-04-09
Add files via upload

veertobb:


nikgapps:
@veertobb the config file must be part of 10 or 11 folder for the build process to understand for which android version we need the gapps. please move the config file to appropriate version and send the pull request again

PR #57, opened 2021-04-12
10somil

somilkhan:


10/somil.config
```diff
@@ -0,0 +1,146 @@
...
+GoogleCalendar=1
+MarkupGoogle=0
+GoogleWallpaper=0
+GoogleFeedback=0
+GooglePartnerSetup=0
+GoogleSounds=0
+AndroidDevicePolicy=0
+GoogleKeep=1
+Books=0
+Assistant=1
```
DrAmoled, line 122: This line will include non functional assistant shortcut because you exclude Velvet

somil.config
```diff
@@ -0,0 +1,146 @@
+# NikGapps configuration file
```
DrAmoled, line 1: Delete this config , only put config in /11 or /10

DrAmoled (changes requested):
Please adjust according comments I put inside your configs

PR #58, opened 2021-04-12
merge this commit

krakeen25:
config file cummit merge pls thanks


DrAmoled:
There is no need to use excessive language.

PR #65, opened 2021-04-13
Nikgappsghidoraha10

angelawhitelub:


Ghidorah.config
```diff
@@ -0,0 +1,146 @@
+# NikGapps configuration file
```
DrAmoled, line 1: Delete this config and only use the one in /11 or /10

DrAmoled (changes requested):
Please delete other config which not in /11 or /10

PR #67, opened 2021-04-14
jey7r.config

jey7r:


nikhilmenghani:
@jey7r don't modify the existing config file. Create your own and then send a pull request.

PR #112, opened 2021-04-28
Google Go Plus Gapps

custosoft:


nikhilmenghani:
@custosoft google go packages are not part of custom builds yet. It's something that we're working on. I will keep you updated when it's ready.

PR #139, opened 2021-05-17
Yasundram patch 1

Yasundram:


nikhilmenghani:
@Yasundram you need to upload your files in either 11 or 10 folder, not outside of it.

Also you need to rename the config file to whatever name you want your custom build to be called. Please change the config file name from nikgapps to a different name.

PR #199, opened 2021-06-03
New build please automagical overlord

stylemessiah:
The fun police are bloating too many ROMS with themes etc, and ruining the users
ability to flash whatever damn size g-apps they want. Dark days, does anyone remember
when you could flash what you want.

Now its less than Omni on some ROMs

*sigh*


nikhilmenghani:
I feel ya

PR #310, opened 2021-07-11
GGM

ggm3000:


nikhilmenghani:
@ggm3000 you don't have to put the whole file name in your config file. You can just name it ggm.config and it will automatically generate the whole name.

ggm3000:
@nikhilmenghani 

Ohh ok
This is my first time using GitHub
Thanks man

PR #366, opened 2021-07-11
Create Inferno1931e

MTadakod:


nikgapps:

- 11/Inferno1931e doesn't have .config extension, we only accept config files!

PR #481, opened 2021-07-13
icecandy-pravinjr-gapps

pravinjagadish:
Hey Nik, thanks for all your hardwork


nikgapps:

- 11/icecandy-pravinjr-gapps.config.txt doesn't have .config extension, we only accept config files!

PR #582, opened 2021-07-14
deep

aakashdeepB:


nikgapps:

- deep.config must be part of Android Version folder, not outside of it!

PR #1022, opened 2021-08-02
build #9

aryaman4real:


nikgapps:

- 11/logic-basic doesn't have .config extension, we only accept config files!

PR #1262, opened 2021-08-14
Real Core

rpy827:


nikgapps:

- Lospack.config must be part of Android Version folder, not outside of it!

PR #1294, opened 2021-08-15
creating gapps package again without go-apps

jaichiranjeeva:


jaichiranjeeva:
@nikhilmenghani looks like latest version of nik.config i.e., v17 having a small issue of building go-apps if corego is set to 0 even , maybe ">>" missing in config for go apps

PR #1686, opened 2021-09-18
Devil's gapps

ToxicPBX1:
Thx👍


nikhilmenghani:
@ToxicPBX1 why would you add a smiley in the name of your config? Such changes only creates problems while uploading the zip to sourceforge.net. Make sure you always use text only in the name.

ToxicPBX1:
Ok next time I will .

PR #1719, opened 2021-09-19
keyoke87_gapps_new_190921

keyoke87:
resubmit. thanks nik


nikgapps:

- 11/keyoke_gapps_config doesn't have .config extension, we only accept config files!

PR #1794, opened 2021-09-25
Create goshawk22-25-09-minimal

goshawk22:


nikgapps:

- 11/goshawk22-25-09-minimal doesn't have .config extension, we only accept config files!

PR #1907, opened 2021-10-10
Add non-aplhanumberic configs

nikgapps:


nikgapps:

- 10/J̰̃ïɳ͡x๛.config is not aphanumeric, make sure the name of config file is between A-Z and 0-9
- 10/😈devilsgapps.config is not aphanumeric, make sure the name of config file is between A-Z and 0-9

PR #1947, opened 2021-10-13
lan2

lanvall:


nikgapps:

- 11/lan_2.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 any symbols including (, - ' . # ! *) are not accepted in the name

PR #1955, opened 2021-10-14
gapps-alexxx-v11-069

Alexxx069:
Hey Nik, Thanks for all the hardwork.


nikgapps:

- 10/my-gapps-v10-0069.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 any symbols including (, - ' . # ! *) are not accepted in the name
- 11/my-gapps-v11-0069.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 any symbols including (, - ' . # ! *) are not accepted in the name

PR #1967, opened 2021-10-15
envy-levi-gapps

gh-levi18:
Will Chrome be fixed?


nikgapps:

- 11/Levi-gapps.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 any symbols including (, - ' . # ! *) are not accepted in the name

PR #2002, opened 2021-10-18
Omni+kok

koktebel5:
v17 => v18


nikgapps:

- 11/Omni+kok.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

PR #2117, opened 2021-10-31
magmaa android 11 gappss

magmareloaded:
thanks nikhil


nikgapps:

- 11/nikgapps(1).config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name
- nikgapps(1).config must be part of Android Version folder, not outside of it!
- nikgapps(1).config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

PR #2127, opened 2021-11-01
LOS18_1-StockGapps-Fajita-CR7

SwitU7Ronald:


nikgapps:

- 11/los18.1-fajita-cr7.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

PR #2151, opened 2021-11-05
Coregappsbynajeeb

mohdnajeeb333:


nikgapps:

- 11/mohdnajeeb333 doesn't have .config extension, we only accept config files!
- Gapps123.config must be part of Android Version folder, not outside of it!

PR #2601, opened 2022-01-15
added new config

rajeshrvds:
added yennifer config


nikgapps:

- 11/yennifer.txt doesn't have .config extension, we only accept config files!

PR #2628, opened 2022-01-18
HavocOSOP8-gapps

androidenthusiast101:
ty for the gapps


androidenthusiast101:
thank you so much it works! no issues! :)

PR #2681, opened 2022-01-23
Create InfernoA12

Alex727863:


nikgapps:

- 12/InfernoA12 doesn't have .config extension, we only accept config files!

PR #2755, opened 2022-02-02
shotg1

shotg1:


nikgapps:

- 12/shotg1 doesn't have .config extension, we only accept config files!

PR #2936, opened 2022-02-25
Bhim new gapps

BHIMESH9121:
Thanks for the support


nikgapps:

- 11/bhim12 .config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name
- 12/bhim12 .config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

BHIMESH9121:
Thanks

PR #3256, opened 2022-04-07
pablomg29-gapps-android-12.1

pablomg29:
Hey Nik, thank for all the hardwork


nikgapps:

- 12/pablomg29-gapps-12.1.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

PR #3307, opened 2022-04-12
datnx82-gapps

xuandat09:
Hi Nik, due to the FC of pixel launcher. Please make new gapps without pixel launcher. Thanks for your hard works!


nikgapps:

- datnx82-gapps.config must be part of Android Version folder, not outside of it!

PR #3324, opened 2022-04-13
Updated gapps with new v21 config

mdbelayethossen:
Thanks for the hard work


nikgapps:

- 11/belayet_custom.config.txt doesn't have .config extension, we only accept config files!
- 11/belayet_custom.config.txt is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name
- Cannot merge the changes automatically since 11/belayet_custom.config.txt is either modified or removed, Wait for someone to manually review!

PR #3444, opened 2022-04-25
rohitfirst12.1gapps

rohit5585:
hey nik thanks for the work u doin man.


nikgapps:

- 12.1/rohit-first-12.1-gapps.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

rohit5585:
is it good now?

PR #3459, opened 2022-04-26
rohitgapps12.1

rohit5585:
thanks nik


nikgapps:

- 12.1/rohit-first-12.1-gapps.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name
- rohitgapps.config must be part of Android Version folder, not outside of it!

PR #3531, opened 2022-05-03
ADR

AlbiDR:
Modified Basic package to include and replace some other Stock apps


AlbiDR:
This .config has '-1' instead of '0'. I hope it will work despite that.

PR #3613, opened 2022-05-15
minh-gapps

pn-minh:


nikgapps:

- minh.config must be part of Android Version folder, not outside of it!

PR #3916, opened 2022-06-30
added_dialer

doggo-87:


nikgapps:

- maw.config must be part of Android Version folder, not outside of it!

PR #3950, opened 2022-07-03
SavanthGC

GurucharanSavanth:
THANK YOU FOR UR WORK


nikgapps:

- 12.1/SavanthGCconfig doesn't have .config extension, we only accept config files!

PR #4020, opened 2022-07-11
omni-extra-gapps

2217dks:


nikgapps:

- 12.1 doesn't have .config extension, we only accept config files!
- 12.1 must be part of Android Version folder, not outside of it!
- 12.1/placeholder doesn't have .config extension, we only accept config files!
- Cannot merge the changes automatically since 12.1/placeholder is either modified or removed, Wait for someone to manually review!

PR #4146, opened 2022-07-28
Core+Gboard

rpy827:


nikgapps:

- 11/placeholder doesn't have .config extension, we only accept config files!
- Cannot merge the changes automatically since 11/placeholder is either modified or removed, Wait for someone to manually review!
- 12.1/Core+Gboard.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) are not accepted in the name

PR #4199, opened 2022-08-06
theprophec-gapps

ThePropheC007:
Hi Nik, you doing really hardwork, thanks for the custom gapps in advance.


ThePropheC007:
there are bugs in this one, Data Transfer tool crashes and unable to pick up call

PR #4254, opened 2022-08-16
PiX_GApps-CLF-v1

pixincreate:
Update to latest build

@nikhilmenghani, are Config builds based on Canary or Stable builds?


pixincreate:
> It's Canary. Always up to date!

Glad to hear that! updated to the latest build today :D

PR #4295, opened 2022-08-21
maxes27

Maxes27:
Hey nik, thanks a lot for the hardwork!


nikgapps:

- 13/maxes27.config must be part of Android Version folder, not outside of it!
- 13/placeholder doesn't have .config extension, we only accept config files!
- 13/placeholder must be part of Android Version folder, not outside of it!

PR #4391, opened 2022-09-05
pixgapps

pixincreate:


pixincreate:
@nikhilmenghani , where are the files? It's unavailable. And I believe, it's being re-directed to build folder

nikhilmenghani:
@pixincreate should be fixed now. If you don't see your build, send another pull request, it will work now. There was an issue with upload which is why previous config build may or may not be up.

PR #4975, opened 2022-11-28
LiteGapps

SangeethSujith:
I removed the extra configs that was in the A13 folder and only kept the config edited by me. Or should i keep the rest of the configs?


nikhilmenghani:
@SangeethSujith You're not supposed to remove anything, just add your config files. If you haven't watched the tutorial video, I suggest you watch it. It has clear instructions and will save your time.

SangeethSujith:
I used to to this before. When i looked inside the a13 folder saw some additional config and they looked like duplicates, So i removed them to avoid any other extra gapps build. Anyway thanks for your fantastic work

PR #5019, opened 2022-12-05
My updated config for RiceDroid

NoName-exe:


nikgapps:

- 13/noname-exe-ricedroid doesn't have .config extension, we only accept config files!

NoName-exe:
What a freaking idiot!!

PR #5072, opened 2022-12-12
tk2p-gapps

tk2p:


nikgapps:

- tk2p.config must be part of Android Version folder, not outside of it!

PR #5087, opened 2022-12-14
axm12.1core

ambanijohnny:


nikgapps:

- 12.1/axm12.1core.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters

PR #5145, opened 2022-12-18
Custom gapps build

auditore9:


nikgapps:

- 12.1/auditore.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 23
- 12/auditore.config is on version 20 which is not the latest version of NikGapps Config, please update the config file to version 23

PR #5185, opened 2022-12-19
bqnvi-vks-bhp-gapps

kumaratv:
Hello Nik,
This is the first time using your service and I am very happy to try this.
Thank you very much for your efforts to this possible.
bqnvi


nikgapps:

- 13/bqnvi-vks-bhp-gapps.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 23

PR #5340, opened 2023-01-02
YashAwaken13

YASHTHEGREAT35:


nikgapps:

- 11/YASHGAPPS.config is on version 19 which is not the latest version of NikGapps Config, please update the config file to version 23
- 11/YASHGAPPS2.config is on version 19 which is not the latest version of NikGapps Config, please update the config file to version 23

PR #5462, opened 2023-01-14
LokiTheDevil Gapps 10Q

MrLokiTheDevil:


MrLokiTheDevil:
Thank you for being a developer

nikgapps:

- 10/LokiTheDevil.config.txt doesn't have .config extension, we only accept config files!
- 10/LokiTheDevil.config.txt is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters

MrLokiTheDevil:
Sorry for uploading .txt file
i forgot to check

PR #5498, opened 2023-01-16
Redmi 5 with Go

54b1d:


nikgapps:

- 13/Redmi5GbTtsAssistGo.txt doesn't have .config extension, we only accept config files!

PR #5510, opened 2023-01-16
huh

alix1383:


nikgapps:

- 13/senpaiapps_T.conf doesn't have .config extension, we only accept config files!
- 13/senpaiapps_T_opt.conf doesn't have .config extension, we only accept config files!

PR #5531, opened 2023-01-18
SayaGapps-minimal

MikeSaya1:


nikgapps:

- All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome

PR #5552, opened 2023-01-20
custom gapps

nmayorga092:


nikgapps:

- 13/nmayorga doesn't have .config extension, we only accept config files!

nmayorga092:
Fixed

PR #5567, opened 2023-01-20
ngapps

shishir266:


nikhilmenghani:
@shishir266 kindly use latest version of nikgapps.config that is v23 to build a custom gapps build.

PR #5603, opened 2023-01-24
Add files via upload

Stefanowall:


nikgapps:

- stefanowall.config must be part of Android Version folder, not outside of it!
- All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #5606, opened 2023-01-24
Add files via upload

Stefanowall:


Stefanowall:
Thank you very much for your excellent work! I flashed the gapps package
made by myself (🤣🤣🤣🤣) and it works fine. I found only one issue, not
experienced before: now, when I use google dialer, a pop up appears which
says that my phone is not suited for that dialer and that calls may be
problematic. Each and every call I majey. Quite annoying, though I can say
that everything works, in reality. If you know something about it, please
tell me😙. Thank you again!

Stefano

Il mar 24 gen 2023, 10:14 Nikhil Menghani ***@***.***> ha
scritto:

PR #5730, opened 2023-02-01
SinisterGapps

vivekvijayan20:


nikgapps:

- 13/SinisterGapps.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24

PR #5743, opened 2023-02-01
Create VOS

FARDUSGTB:
THX


nikgapps:

- 13/VOS doesn't have .config extension, we only accept config files!

PR #5797, opened 2023-02-04
Blue 

Flamo0:
Thankyou


nikgapps:

- 11/Blue.config is on version 19 which is not the latest version of NikGapps Config, please update the config file to version 24
- 11/Flamo.config is on version 17 which is not the latest version of NikGapps Config, please update the config file to version 24
- 12.1/blue.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 24
- 12/Flamo.config is on version 17 which is not the latest version of NikGapps Config, please update the config file to version 24

nikgapps:

PR #5799, opened 2023-02-04
Thanks for build :)

petrzmax:


nikgapps:

- 11/petrzmaxFullLightV11.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 24

petrzmax:
Fixed

PR #5808, opened 2023-02-05
ngoudar-gapps

ngoudar1729:
Thanks for build


nikgapps:

- 10/ngoudar.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24
- 11/ngoudar.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24
- 12.1/ngoudar.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24
- 13/ngoudar.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24

PR #5849, opened 2023-02-08
Lineageos-20-Final-13-24

Aryoper:


nikgapps:

- 13/Lineageos-20-Final-13.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24
- 13/Lineageos-20-Final-Pixel-Launcher2.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24
- 13/Lineageos-20-Final.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 24

PR #5915, opened 2023-02-14
N5672288372772

Offici5l:
tnks


nikgapps:

- N5672288372772.config must be part of Android Version folder, not outside of it!
- Update N5672288372772.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #5938, opened 2023-02-16
Full-P

HARDWELL9191:


nikgapps:

- Update 13/Full-P.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #5971, opened 2023-02-18
Roe-gapps

Ragdox:
Thanks Nik


nikgapps:

- Update 13/Roe.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 13/Roe.config, You cannot enable both Core and Core Go, you can only enable one of them

Ragdox:
Thanks

PR #5973, opened 2023-02-18
Custom config

Wossgebber:
Thanks in advance.


nikgapps:

- 12.1/wossgebberSLcustomgapps doesn't have .config extension, we only accept config files!

PR #5982, opened 2023-02-18
Add silverethical config for android 13

Silverethical:
Appreciate your work.


nikgapps:

- Update 13/silverethical.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #5999, opened 2023-02-19
twoface-v3

TwoFace0:
thanks nik


nikgapps:

- 13/twoface-v3.config is on version 24 which is not the latest version of NikGapps Config, please update the config file to version 25

PR #6022, opened 2023-02-20
Spark mml no Velvet

snascs:


nikgapps:

- 13/snascs-a13-spark-mml-no-velvet.config is on version 24 which is not the latest version of NikGapps Config, please update the config file to version 25

PR #6028, opened 2023-02-20
Snascs a13 v25

snascs:


nikgapps:

- Update 13/snascs-a13.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- snascs-a13.config must be part of Android Version folder, not outside of it!
- Update snascs-a13.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #6046, opened 2023-02-22
RishaGo

SilverRisha:


nikgapps:

- RishaGo.config must be part of Android Version folder, not outside of it!

PR #6059, opened 2023-02-23
back to nickgapps :)

sssira:


nikgapps:

- 12.1/sssira-semifull-edit.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 25

PR #6113, opened 2023-02-26
ASZ002

djang0j0e:


nikgapps:

- Update 13/nikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them
- ASZcust.config must be part of Android Version folder, not outside of it!

PR #6118, opened 2023-02-27
Add files via upload

shironamiyumi:
No pixel launcher


nikgapps:

- 13/no pixel launcher .config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Update 13/no pixel launcher .config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6128, opened 2023-02-27
rgood5-gapps

rudra-good:


nikgapps:

- rgood5.config must be part of Android Version folder, not outside of it!

PR #6132, opened 2023-02-27
nikgapps sonu

arvik8660:
thank u nikk


nikgapps:

- 11/ancient-gapps.config is on version 17 which is not the latest version of NikGapps Config, please update the config file to version 25
- 13/nikgapps sonu.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- nikgapps sonu.config must be part of Android Version folder, not outside of it!
- nikgapps sonu.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters

PR #6167, opened 2023-03-02
dk14_nik

DarkKnight1444:


nikgapps:

- Update 13/dk14_nikgapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #6194, opened 2023-03-04
A13 Lite gapps

subhashis-barad:


nikgapps:

- 13/subhashisbarad-13.config is on version 24 which is not the latest version of NikGapps Config, please update the config file to version 25

PR #6210, opened 2023-03-05
pixel-launcher-less

BAYMAX-Maruf:
Thanks


nikgapps:

- Update 13/pixel-launcher-less.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6223, opened 2023-03-06
Bestgapps123

mohdnajeeb333:


nikgapps:

- 11/mohdnajeeb333.config is on version 24 which is not the latest version of NikGapps Config, please update the config file to version 25
- 12.1/mohdnajeeb333.config is on version 24 which is not the latest version of NikGapps Config, please update the config file to version 25
- Update 12.1/mohdnajeeb333.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #6228, opened 2023-03-06
Aegis-37 Gapps

zidan-m41:
Custom Gapps


nikgapps:

- Update 12/aegis-37-gapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 12/aegis-37-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them
- Update 13/aegis-37-gapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 13/aegis-37-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6253, opened 2023-03-09
nikgapps-elzyy

elzyy:
Thanks for the amazing tool nikgapps


nikgapps:

- Update 13/nikgapps-elzyy.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6286, opened 2023-03-11
Aegis-37 New Update

zidan-m41:


nikgapps:

- 12/aegis-37-gapps.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 26
- 13/aegis-37-gapps.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 26

PR #6303, opened 2023-03-11
whoube.nikgapps

whoube:
Hey Nik thanks

### DIFF
--- a/13/whoube.nikgapps.config
+++ b/13/whoube.nikgapps.config
@@ -1,0 +1,135 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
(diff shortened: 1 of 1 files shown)


nikgapps:

- 13/whoube.nikgapps.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Update 13/whoube.nikgapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

whoube:
Resubmitting with correct name format

PR #6315, opened 2023-03-12
ptr

ptr1968:
tnx, only the .config ones


nikgapps:

- Update 13/ptrnikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them
- ptr1968.config.txt doesn't have .config extension, we only accept config files!
- ptr1968.config.txt must be part of Android Version folder, not outside of it!
- ptr1968.config.txt is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters

nikgapps:

PR #6317, opened 2023-03-12
Gapps rispes

RINOIP30:


nikgapps:

- 13/Rispes_core_gapps.bin doesn't have .config extension, we only accept config files!

PR #6319, opened 2023-03-12
MJ-NikGapps-March

Chappie18:


nikgapps:

- 13/MJnikgapps.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 26

PR #6335, opened 2023-03-14
CodeChas3r-gapps

CodeChas3r1:
Hey nik, thank you for your hardwork.

thanks always


nikgapps:

- 13/CodeChas3r.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 26

PR #6350, opened 2023-03-15
a11-Gapps-Fr0ztyy43

Fr0ztyy43:


nikgapps:

- 11/froztyy43.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 26

PR #6360, opened 2023-03-16
q for spartan

subhashis-barad:


nikgapps:

- 10/nikgapps Q.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- 10/q for spartan.txt doesn't have .config extension, we only accept config files!
- 10/q for spartan.txt is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- 13/subhashis_13.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 26

nikgapps:

PR #6368, opened 2023-03-16
T config

Wossgebber:


nikgapps:

- 12.1/wossgebberSLcustom doesn't have .config extension, we only accept config files!

PR #6440, opened 2023-03-24
A12.1

vampent0026:
always thanks to your hardwork niks.


nikgapps:

- Update 12.1/bencena0026-SL.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6487, opened 2023-03-29
Goku-minnu-gaaps

Gokuminnu:
Hy nik, thanks a lot for your hard work


nikgapps:

- Update 12.1/goku-minnu.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6490, opened 2023-03-29
GansGapps-A12L

abdlgni:
Hi


nikgapps:

- Update 12.1/GansGapps-A12L.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6534, opened 2023-04-02
Gapps-without-pixel-launcher-crdroid

Maruf-62783:
Gg

### DIFF
--- a/13/pixel launcher less.config
+++ b/13/pixel launcher less.config
@@ -1,0 +1,133 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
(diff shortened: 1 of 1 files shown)


nikgapps:

- 13/pixel launcher less.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- 13/pixel launcher less.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 26
- Update 13/pixel launcher less.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6580, opened 2023-04-06
MM

mono66985:


nikgapps:

- 13/nikgapps.config is on version 26 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #6588, opened 2023-04-07
Snascs v26.1

snascs:


nikgapps:

- 13/snascs-a13.config is on version 26 which is not the latest version of NikGapps Config, please update the config file to version 27
- Update 13/snascs-a13.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #6605, opened 2023-04-07
Blazkowicz

AndroidEnthusiast:
Thanks Nik


nikgapps:

- Blazkowicz.config must be part of Android Version folder, not outside of it!
- Blazkowicz.config is on version 26 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #6663, opened 2023-04-13
nino1802-gapps

nino1802:
Hay Nik, thx for all the hardwork


nikgapps:

- Update 13/nino1802.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6678, opened 2023-04-15
7929hw-basic-T

7829hw:
Thanks you for your hardworks, nik.


nikgapps:

- Update 13/7929hw-basic-T.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6684, opened 2023-04-15
New Pull Req

XaveriusJuan:


nikgapps:

- 10/CQGApps-AndroidQ.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 27
- 13/CQGApps AndroidT v27.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- 13/CQGApps-AndroidT.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #6754, opened 2023-04-21
linWi3-gapps

Wi3n3:
thx for the hard work


nikgapps:

- Update 13/linWi3.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6758, opened 2023-04-22
Vampi

Vamsi-1925:
Thanks for creating this amazing gapps


nikgapps:

- Update 13/Vamsi.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6769, opened 2023-04-23
Vampi1

Vamsi-1925:
Please check this.
Thank you


nikgapps:

- Update 13/Vamsi.config, You cannot enable both Core and Core Go, you can only enable one of them
- Vamsi.config must be part of Android Version folder, not outside of it!

PR #6849, opened 2023-05-02
Crdroid gapps

Squid1984:


nikgapps:

- 11/Baikal_SG.config is on version 23 which is not the latest version of NikGapps Config, please update the config file to version 27
- Update 11/Baikal_SG.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #6868, opened 2023-05-04
dykey-gapps

DyKey13:
Hi. Thanks for everything! Is great!


nikgapps:

- 12/nikgapps.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 27
- Update 12/nikgapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

DyKey13:
I think is ready now. Updated and Chrome included

PR #6888, opened 2023-05-07
Update config for Prototyped.

Prototyped:
Update config for Prototyped, in particular updating config schema version to 27.


nikgapps:

- Update 10/Prototyped.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #6919, opened 2023-05-10
Add Elite

nikhilmenghani:


nikgapps:

- elite/10/placeholder doesn't have .config extension, we only accept config files!
- elite/11/placeholder doesn't have .config extension, we only accept config files!
- elite/12.1/placeholder doesn't have .config extension, we only accept config files!
- elite/12/placeholder doesn't have .config extension, we only accept config files!
- elite/13/A/mtg.config is on version 25 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #6947, opened 2023-05-15
ikhder-Gapps

ikhder:
Thanks do much nik


nikgapps:

- 12.1/KhderGapps.config is on version 22 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #6960, opened 2023-05-16
saroj-gapps11

saroj-nokia:
hey Nik thanks for your hardwork.


nikgapps:

- Update 11/saroj-gapps11.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7039, opened 2023-05-26
1st try

MdeBovenste:
1st try user build nikgapps


nikgapps:

- Update 13/mdb.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7093, opened 2023-06-03
reducing the space needed for oneplus9 pro

pecha2k2:


pecha2k2:
thanks for sharing this tool

PR #7124, opened 2023-06-04
masr944-tulip

Masr944:
Hey Nik, tks for all hardwork.


nikgapps:

- 10/masr944-v10.0.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Update 10/masr944-v10.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

Masr944:
Tks

PR #7131, opened 2023-06-05
JCGapps

skydeocampo17:


nikgapps:

- 11/JCApps.config is on version 20 which is not the latest version of NikGapps Config, please update the config file to version 27

PR #7169, opened 2023-06-10
SRH_K20P_Gapps

a2zshahriar:
Thank you very much for your valuable support.


nikgapps:

- elite/13/SRH_K20P.config is not a valid path

PR #7178, opened 2023-06-11
micro-nikgapps-withsetupwizard

rosemaryuser:
thanks for creating such a wonderful gapps project


nikgapps:

- 12.1/edited.config is on version 27 which is not the latest version of NikGapps Config, please update the config file to version 28

PR #7199, opened 2023-06-12
update config version

Qiraht:


nikgapps:

- Update 13/qiraht.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7221, opened 2023-06-15
Add files via upload

rico-vz:


nikgapps:

- Update 13/cr-rev-june15.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/cr-rev-june15.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7225, opened 2023-06-15
One more gapps

LukeMech:
XD device setup needed play store not working properly


LukeMech:
Well it works without changes 😅

PR #7231, opened 2023-06-16
Gapps-Upen-crDroid

UpenPou:
Thank You Nik. This is For New Version of crDroid.


nikgapps:

- 13/Gapps-Upen doesn't have .config extension, we only accept config files!
- 13/Gapps-Upen is on version 27 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-T/v28/)and update the config file to version 28

PR #7234, opened 2023-06-17
For-Crdroid-Upen

UpenPou:
I could not upload the file so i went this way. Thank you so much


nikgapps:

- 13/Upen-Crdroid doesn't have .config extension, we only accept config files!
- Update 13/Upen-Crdroid, You cannot enable both Core and Core Go, you can only enable one of them

PR #7239, opened 2023-06-17
Here is my req

ARAIsham:
Please approve


nikgapps:

- Update 13/isham.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7246, opened 2023-06-17
crDroid-op6: Test dropping GoogleTTS

Terminator-J:
Our system partition is too small, but need to install Google app (velvet) as priv-app in order for hotword detection to work correctly.

Try dropping GoogleTTS and hope that it can be installed from Play Store after first boot & still work correctly.

(just doing this as a personal build to test things before moving it to elite build config)


Terminator-J:
Ack, was I supposed to strip the existing "elite" stuff from the config file before reusing it for a personal test build?

PR #7253, opened 2023-06-19
tefa-gapps

mustafahamdydev:
Thnx for the hard work


nikgapps:

- Update 13/tefa.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7257, opened 2023-06-19
Erebit_gapps

tibere86:
Hey Nik, many thanks for your hardwork.


nikgapps:

- Update 13/Erebit_gapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7259, opened 2023-06-19
rakmoparte-gapps

RAM-UNLOK:
Hey Thanks a lot


nikgapps:

- Update 13/rakmoparte.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7268, opened 2023-06-20
AHGapps-20230609-1-v28

arifinhakim87:


nikgapps:

- 10/AHGapps-10-20230609-1-v28 doesn't have .config extension, we only accept config files!
- 12.1/ah-slgapps.config is on version 23 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-SL/v28/)and update the config file to version 28

PR #7366, opened 2023-06-28
yasen-gapps

YasenYanev:
Thx for the hard work!


nikgapps:

- Update 13/yasen-gapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/yasen-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7367, opened 2023-06-28
nikgapps28.06

udittalwar:


nikgapps:

- 12.1/udgapps.config is on version 22 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-SL/v29/)and update the config file to version 29
- Update 12.1/udgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/nikgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/nikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7380, opened 2023-06-28
Update GApps while not including Google Chrome.

Prototyped:
Update GApps while not including Google Chrome.


nikgapps:

- Update 10/Prototyped.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 10/Prototyped.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7421, opened 2023-07-01
LEE-gapps

Mouleesh23:


nikgapps:

- Update 13/LEE.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- 13/mouleesh.config is on version 26 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-T/v29/)and update the config file to version 29

PR #7466, opened 2023-07-06
AddonsEvolutionX-nikgapps

Darrellwibu02:
This is a custom gapps for evolution x The "Addons" Name is Already Remove its should no Problem For that Thanks for your hardwork


nikgapps:

- 13/AddonsEvolutionX-nikgapps.config contains Addon in the name, please avoid using Addon in the name. It conflicts with the official Addon naming convention!

PR #7475, opened 2023-07-07
reuploaded new gapps 

BerseusBerat:


nikgapps:

- 11/a11bero (1).config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Update 11/a11bero (1).config, You cannot enable both Core and Core Go, you can only enable one of them

BerseusBerat:
https://github.com/nikgapps/config/pull/7475/commits/834ee35b3b58ac9d6436f7861bee393e6e8833f7  updated config

PR #7525, opened 2023-07-11
Aegis-37 Gapps

zidan-m41:


nikgapps:

- Update 13/aegis-37-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #7537, opened 2023-07-12
taimen product gapps test2

SunilPrasadRegmi:


nikgapps:

- Update 13/taimen_product_gapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7561, opened 2023-07-13
Custom GApps Package for Munch - Android 13

DeadlyShroud:
Thank you for the service, Nikhil!


nikgapps:

- Update 13/Munch.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7714, opened 2023-07-21
bitra-gapps

RAM-UNLOK:
Thanks Nik you are the best.


nikgapps:

- Update 13/rakmo-bitra.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7717, opened 2023-07-21
BlissGapps 

THEB1GB:
2nd iteration of the custom nikgapps for blissrom vanilla build 
1. This removes the aosp jellyfish browser in blissrom.
2. I hope this will also fix the bug that installs YouTube app even though I didn't allowed it in the config.


nikgapps:

- Update 13/Bliss.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #7736, opened 2023-07-23
myconf-gapps

Angekill2:
thanks for everything, great work!


nikgapps:

- Update 13/myconf.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7802, opened 2023-07-28
crDroidGapps

vivekvijayan20:


nikgapps:

- 13/SinisterGappscrDroid.config1 doesn't have .config extension, we only accept config files!

PR #7877, opened 2023-08-01
losno11

AnishiMK:


nikgapps:

- los11.config must be part of Android Version folder, not outside of it!

PR #7927, opened 2023-08-03
gapps-harshit

harshith3664:
Hey Nik, thanks for all that hard work.


nikgapps:

- Update 13/harshit.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #7969, opened 2023-08-05
Custom Elite Nikgapps package for Mi 9T- AlphaDroid

helliscloser:


nikgapps:

- helliscloser is not allowed to upload to elite/13/alphadroid_davinci/alphadroid.config
- Update elite/13/alphadroid_davinci/alphadroid.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

helliscloser:
Can I skip both though? As the rom might already ship setup wizard. I would only need to ship google restore.

PR #7971, opened 2023-08-05
DroidX-UI

NRanjan-17:


nikgapps:

- 13/NikGapps-DroidX-UI-SE-arm64-13-20230805-signed.zip doesn't have .config extension, we only accept config files!

PR #8004, opened 2023-08-07
Sureshgorai25 NikGapps

Sureshgorai25:
Hey Nik, Thanks for the Gapps


nikgapps:

- Update 13/Suresh25.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8013, opened 2023-08-08
ak17-arm64-13-20230808

aghilesh17:


nikgapps:

- Update 13/ak17-arm64-13-20230808.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8022, opened 2023-08-09
SinisterGapps

vivekvijayan20:


nikgapps:

- Update elite/13/Sinister/SinisterGapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8139, opened 2023-08-17
P1x3l-H4rdDr1v3

BenHidalgo:
Pixelite


nikgapps:

- Update 13/H4rdDr1v3.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8205, opened 2023-08-22
merge kvarxn config

kvxrun:


nikgapps:

- 11/kvxrun.config is missing version information, please use latest version of NikGapps Config i.e. 30
- Update 11/kvxrun.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/kvarxn.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8295, opened 2023-08-31
DarrellOkkotsu3040-nikgapps

DarrellOkkotsu3040:
There's been a long time but btw can you do this for me thanks for your hardwork


nikgapps:

- Update 13/DarrellOkkotsu3040-nikgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8305, opened 2023-09-03
Akash

akashsriramganapathy:


nikgapps:

- Update 13/aksriram-nikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #8367, opened 2023-09-11
Q update

vayuponeee:


nikgapps:

- 10/Ten-Verse.txt doesn't have .config extension, we only accept config files!
- Update 10/Ten-Verse.txt, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #8374, opened 2023-09-11
gapps for android 13

jpwers:


nikgapps:

- Update 13/Jgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8637, opened 2023-10-11
Update (fixed)

Wossgebber:


nikgapps:

- elite/13/Wossgebber/LOS/LineageOSTcustomgapps.config is not a valid path, make sure your elite config is in elite/<android_version>/<elite_folder_name>
- Update elite/13/Wossgebber/wossgebberTcustomgapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #8700, opened 2023-10-17
gapps-a13

phantan123:
Thanks for your work Nik


nikgapps:

- Update 13/phantan123.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

phantan123:
Here is the updated config

PR #8711, opened 2023-10-18
MikuX Android 10-2

MikuX-Dev:
Here we go bruh


nikgapps:

- Update 10/MikuX-nikgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8788, opened 2023-10-26
peter4vn-f3-cr-13

peter4vn:
Great work. Thank you.


nikgapps:

- Update 13/peter4vn.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 13/peter4vn.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #8859, opened 2023-11-09
my a14 config

vytska69:


nikgapps:

- 14/placeholder doesn't have .config extension, we only accept config files!

PR #8860, opened 2023-11-09
my a14 gapps

vytska69:


nikhilmenghani:
It'll be added to queue. It won't get built. Android 14 support will be added once the builds are marked as stable.

PR #8861, opened 2023-11-09
A14 First Build

invisible789:
Thanks


nikhilmenghani:
It'll be added to queue. It won't get built. Android 14 support will be added once the builds are marked as stable.

invisible789:
> It'll be added to queue. It won't get built. Android 14 support will be added once the builds are marked as stable.

Ok, thanks for informing, looking forward to future developments

PR #8875, opened 2023-11-11
dick_papps

dcoxii1979:
Thank you nik


nikgapps:

- Update 13/dick_papps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8918, opened 2023-11-13
Custom config for gapps

samio5:
Thanks Nikhil for your hard work.


nikgapps:

- Update 14/ocs-nikgapps.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8953, opened 2023-11-17
crdroid10-A14

YudaAditya:
Hey Nik thanks for all the hardwork


nikgapps:

- Update 14/crdroid10-A14.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #8969, opened 2023-11-19
built-for-crdroid-a14-miatoll

parasgargexx20:
hey thanks for ur hardwork


nikgapps:

- Update 14/perforiented.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 14/perforiented.config, You cannot enable both Core and Core Go, you can only enable one of them
- nikgapps.config must be part of Android Version folder, not outside of it!

PR #8971, opened 2023-11-19
Lptje

LptjeGitHub:
R2. Practice makes Perfect.Thanx for ur patience Bro


nikgapps:

- Update 11/2023-11-19_Gapps-T_Hu_AGS2-L09_v30_2023-07-13-R2.config, You cannot enable both Core and Core Go, you can only enable one of them
- Update 11/2023-11-19_Gapps-T_Hu_AGS2-L09_v30_2023-07-13.config, You cannot enable both Core and Core Go, you can only enable one of them
- Lptje.config must be part of Android Version folder, not outside of it!

PR #8972, opened 2023-11-19
LptjeHuAgs2-L09

LptjeGitHub:
<3


nikgapps:

- Update 11/Lptje.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #8991, opened 2023-11-20
rey-gapps

comglorerate:
Thank you very much for your work


nikgapps:

- Update 13/rey.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9047, opened 2023-11-27
droidwina13.config

Droidwin:
This GApps is created for Android 13 and has most of the apps from the Basic packages. Thanks, Nik for this amazing Gapps package!


nikgapps:

- droidwina13.config must be part of Android Version folder, not outside of it!

PR #9059, opened 2023-11-29
mtcpatcher-gapps

mtcpatcher:
Hello Sir, I'm big fan of you. Thanks for your hardwork.


nikgapps:

- Update 11/mtcpatcher-gapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 11/mtcpatcher-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #9076, opened 2023-12-01
Minor changes to avoid some flashing errors.

jason-eu:
Changed installation folder back to default.
Descriptions.

### DIFF
--- a/14/CrDroid10-A14_jason-eu.config
+++ b/14/CrDroid10-A14_jason-eu.config
@@ -1,0 +1,157 @@
+# CrDroid10.0-A14_jason-eu.config is a NikGApps config file customised for my personal use, on POCO F2 Pro/Xiaomi Redmi K30 Pro/lmi.
+# Intended for use with "NikGapps-full-arm64-14-20231107-signed.zip, of MD5 ; 68e8fa50efabbda765a4e2258932ea66.
+# Intended for use with CrDroid 10.0 (Android 14) ROM, maintained by 'raystef66'.
+# If unsure about a config, don't make changes to it, or comment it by adding '#' before it.
+# visit https://nikgapps.com/misc/2022/02/22/NikGapps-Config.html to read everything about nikgapps.
+# visit https://nikgapps.com/misc/2021/08/24/Packages-Definition.html to read descriptions.
+
+AndroidVersion=14
+Version=30
+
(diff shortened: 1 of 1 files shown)


nikgapps:

- Update 14/CrDroid10-A14_jason-eu.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

jason-eu:
Thank you, but I have successfully flashed the config on my device with ;

```
# Set GoogleChrome=0 if you want to skip installing all packages belonging to GoogleChrome Package.
GoogleChrome=1
>>GoogleChrome=0
>>WebViewGoogle=1
>>TrichromeLibrary=1
```

So I don't really understand why I am getting the conflict.

Is it because main requires it to be all on or all off, even though I can customise it to not install the Chrome browser itself ?

jason-eu:
I don't get it mate, what's the point of configuration then ?

nikgapps:

- Update 14/CrDroid10-A14_jason-eu.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

nikhilmenghani:
> I don't get it mate, what's the point of configuration then ?

The configuration let's you choose the packages independently, but when you build a zip out of it, it gets uploaded as a package and anyone can flash that blindly (surprisingly people do that), there are some rules that as per recommendations are controlled to ensure users don't run into known issues. Other rules are around Core and CoreGo packages (both can't be enabled because you'd just be duplicating stuff when you will NEVER intend to install both at the same time) same goes with SetupWizard and PixelSetupWizard.

I can make an exception if you will want to go ahead and disable GoogleChrome, send another pull request and I will merge it.

jason-eu:
The bot/auto responder made me squint, lol.

That's the thing, Nikhil, I'm not an Elite member yet, I'm working towards that.
I thought that in order to get a package published on the site/sourceforge/telegram User Builds, I'd have to be an Elite Member first.
So I thought okay, I'll test a config and upload it, share it in the channel/group for my device for feedback/issues/bugs, then later on follow one of the pathways to becoming an Elite member.

A while ago, I learned that "/system_ext" can only eat so much, and my system was quite upset with CoreGo and Core overlapping.

Yes, I agree, there are many members of the groups I'm in and my own group that make some…questionable decisions.

That line about CoreGo should not be used with Core in the config should be in capitals 😄 .

nikhilmenghani:
yea, tag me when you send another pull request, I will merge it.

PR #9089, opened 2023-12-01
Delete 14/CrDroid10-A14_jason-eu.config

jason-eu:
It needs removal, as I cannot in good faith recommend it for the purpose it was intended, as there are certain changes that need to be made which are restricted from being merged.

If that changes, I will make a new one, until then I'll fork it.

I do not want anyone to go to the official NikGApps repo and find it, then later make reference to the NikGApps maintainers or myself about it when I have no intention of continuing it there for the moment.


jason-eu:
Thank you, Nikhil, the bot or auto responder was making me worry.

I understand as much, but one of the last changes I was trying to make was to use it like this ;

```
# Set GoogleChrome=0 if you want to skip installing all packages belonging to GoogleChrome Package
GoogleChrome=1
>>GoogleChrome=0
>>WebViewGoogle=1
>>TrichromeLibrary=1
```

This is how I've used the config file for NikGApps for some time without issue on Android 11 (I think), 12, 13 and now 14.

I just couldn't understand why it would block me from merging a change in which I only excluded the browser component.
Which is a config that works fine (for me) using the stock ROM browser or something else like Brave/Bromite/Cromite.

nikhilmenghani:
Send another pull request and i will merge it.

jason-eu:
Thank you, I saw your message on #9076 as well.

These days my device is having some issues with an A12 unofficial TWRP recovery that was adapted for A13 that was re-adapted for A14.

So, I have to clean flash again then dirty flash the nikgapps.config I made, with some more changes, to test it because I don't want to put it "out there" before doing that.

In theory, all should be fine and I can chance it with my own device but I'm afraid that someone else will not have the same chances.
And then the one repo I have on GitHub gets a bad rep lol.

I'll have to flash tomorrow afternoon after class to verify all is well then submit the PR and @ you to indicate that it's the one with the Chrome browser exclusion change.

PR #9219, opened 2023-12-11
a71 a14 build

lebbi97:
hey thanks for effort !


nikgapps:

- Update 14/a71.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

lebbi97:
i have updated the file and i think no more conflict, disabled setup wizard and only enabled pixel setup wizard. again thanks for efforts

PR #9255, opened 2023-12-17
NikGapps-RURU-12172023

Raaaawr357:
Please and thanks for all the hardwork NikGapps team! <3


Raaaawr357:
Hi. can I update the configs i made? I tried flashing it but its failing because it says "not enough space" I will attach here the file. All I want is like nikcore setup with added gboard and setupwizard. Thanks

PR #9345, opened 2023-12-30
CrDroid-A14-ex-PL-yt

Maruf-62783:


nikgapps:

- Update 14/CrDroid-A14-without-ex-pl.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- CrDroid-A14-without-ex-pl.config must be part of Android Version folder, not outside of it!

PR #9347, opened 2023-12-30
CrDroid-A14-without-ex-pl

Maruf-62783:


nikgapps:

- Update 14/CrDroid-A14-without-ex-pl.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9353, opened 2023-12-30
merlinx-a13-gapps

songogit:
Thx Nik for the hard work!


nikgapps:

- Update 13/merlinx-a13.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9373, opened 2024-01-01
newbliss_14.config

Annette7582:
This GApps is created for Android 14 build and has most of the apps from the Basic packages itself


Annette7582:
Hey Nikk Thanks For Gapps

PR #9380, opened 2024-01-01
nemesgapps_14.config

Sahib75:
Hey Nik, thanks for your GApps!


nikgapps:

- Update 14/nemesgapps_14.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #9468, opened 2024-01-12
LeanzazzyCustomNikGappsA14

Leanzazzy:
Added only LeanzazzyCustomNikGappsA14 and removed older versions.


nikgapps:

- Update 14/LeanzazzyCustomNikGappsA14.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9489, opened 2024-01-13
config: Release naups-GAPPS for Android 13 (13-01-2023)

naups:
merge


nikgapps:

- Update 13/naups.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9567, opened 2024-01-23
420rom_Full_Gapps_Config

abun880007:
Initial config for 420rom


nikgapps:

- Update 14/420rom_full.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9656, opened 2024-01-29
Sweaty Balls

Drsexo:


nikgapps:

- 11/balls.config is on version 31 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-R/v32/) and update the config file to version 32.

Mismatched config version Sucks, I know!

PR #9728, opened 2024-02-05
HackOs14_launcheradded_gapps

Hack-0s:
Thanks Nik


nikgapps:

- Update 14/HackOs14_launcheradded.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9767, opened 2024-02-06
mrmeqi

mrmeqi:
hey nik thanks for all the hardwork


nikgapps:

- Update 13/mrmeqi.config, You cannot enable both Core and Core Go, you can only enable one of them

mrmeqi:
ok. thanks :)

PR #9774, opened 2024-02-07
white _torch_gappsA14.config

Kvijay199428:
Minimal gapps


nikgapps:

- Update 14/gapps.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

PR #9800, opened 2024-02-08
A14cnfg

AAyar94:


nikgapps:

- .DS_Store must be part of Android Version folder, not outside of it!
- Update 14/aayar94_udc.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9943, opened 2024-02-20
a14-20240220-ashutosh_ishere

ashutosh-ishere:
thanx nik.

### DIFF
--- a/12.1/ashutosh-ishere-nikgapps.config
+++ b/12.1/ashutosh-ishere-nikgapps.config
@@ -1,0 +1,137 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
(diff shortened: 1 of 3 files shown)


nikgapps:

- 12.1/ashutosh-ishere-nikgapps.config is on version 32 which is not the latest version of NikGapps Config, please download the [latest version](https://sourceforge.net/projects/nikgapps/files/Releases/Config/NikGapps-SL/v33/) and update the config file to version 33.
- 14/vos3.2-ashutosh.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters

Mismatched config version Sucks, I know! but you can become [NikGapps Elite](https://nikgapps.com/misc/2023/06/08/NikGapps-Elites.html) and get rid of this hassle. The config will be auto upgraded to latest version.

PR #9973, opened 2024-02-23
update basic-ionut and full-ionut

ionutgherman:


nikgapps:

- Update 14/basic-ionut.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them
- Update 14/full-ionut.config, You cannot enable both Pixel Setup Wizard and Setup Wizard, you can only enable one of them

PR #9995, opened 2024-02-24
Shreets-package

ShreetGupta:


nikgapps:

- Update 10/Shreets-package.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10005, opened 2024-02-25
MKGapps

doomplex:
Cheers


nikgapps:

- Update 14/MKGapps.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10011, opened 2024-02-25
Create Kr_nikgaaps

Kr0069:
Thanks for your hard work..


nikgapps:

- 14/Kr_nikgaaps doesn't have .config extension, we only accept config files!
- Update 14/Kr_nikgaaps, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.

Kr0069 (commented):
I have done the work now

PR #10029, opened 2024-02-29
malaki14pixel

malaki360studio:


nikgapps:

- Update 14/malaki14pixel.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10073, opened 2024-03-05
updated package

Wossgebber:


nikgapps:

- elite/13/Wossgebber/Custom config - T.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Update elite/13/Wossgebber/Custom config - T.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.
- Update elite/13/Wossgebber/Custom config - T.config, You cannot enable both Core and Core Go, you can only enable one of them
- elite/14/Wossgebber/elite/14/Wossgebber/wossgebberUcustomgapps.config is not a valid path, make sure your elite config is in elite/<android_version>/<elite_folder_name>

PR #10075, opened 2024-03-06
updated package

Wossgebber:


nikgapps:

- elite/14/Wossgebber/elite/wossgebberUcustomgapps.config is not a valid path, make sure your elite config is in elite/<android_version>/<elite_folder_name>

PR #10122, opened 2024-03-09
MaxRone7089-gapps

MaxRone7089:
Hi Nikhil, thanks for all these hardworks


nikgapps:

- Update 11/MaxRone7089.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10123, opened 2024-03-09
Create vivalenta.config

vivalenta:


nikgapps:

- vivalenta is not allowed to upload to elite/14/vivalenta/vivalenta.config
- Update elite/14/vivalenta/vivalenta.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10236, opened 2024-03-22
android14.config

MakeMeCool:
This GApps is created for Android 14 and has most of the apps from the Basic packages.


nikgapps:

- Update 14/android14.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10519, opened 2024-04-24
new minimal gapps a14

frankh93:
please and thank you


nikgapps:

- Update 14/PJRR-A14.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10649, opened 2024-05-05
jackssmitgapps

jackssmit:


nikgapps:

- Update 14/jackssmitgapps.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.
- Update 14/jackssmitgapps.config, You cannot enable both Core and Core Go, you can only enable one of them
- jackupdate.config must be part of Android Version folder, not outside of it!

PR #10688, opened 2024-05-11
Jagermeister-gapps

jagermeisterdealer:
Thanks Nik for all the hardwork :heart:


nikgapps:

- Update 14/jagermeisterdealer.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10705, opened 2024-05-12
dorikly-android-gapps

dorikly:
Thank you nik

### DIFF
--- a/14/dorikly-android.config
+++ b/14/dorikly-android.config
@@ -1,0 +1,146 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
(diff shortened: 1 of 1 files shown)


dorikly:
Thank you nik

nikgapps:

- Update 14/dorikly-android.config, All the packages under Google Chrome needs to be enabled, you cannot disable any of the packages under Google Chrome.
- Update 14/dorikly-android.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.
- Update 14/dorikly-android.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10750, opened 2024-05-16
selfgapps

pawan1710:
tnx for your hard work

### DIFF
--- a/14/forself.config
+++ b/14/forself.config
@@ -1,0 +1,145 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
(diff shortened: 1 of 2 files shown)


pawan1710:
thnx for the hard work

nikgapps:

- Update 14/forself.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.
- Update 14/forself.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10776, opened 2024-05-17
VinNikGapps37

VinayVG3:


nikgapps:

- Cannot merge the changes automatically since status of 13/placeholder is removed, kindly start fresh with forking the repository again!
- 13/placeholder doesn't have .config extension, we only accept config files!
- Update 14/vinnikgapps37.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10804, opened 2024-05-20
alash gapps

akash2784:


nikgapps:

- 14/akash2784. config doesn't have .config extension, we only accept config files!

PR #10870, opened 2024-05-28
0x-gapps

0xSoul24:
yes


nikgapps:

- 0xSoul24 is not allowed to upload to elite/14/crdroid-official/crdroid-official.config
- Update elite/14/crdroid-official/crdroid-official.config: You cannot enable both PixelSetupWizard and SetupWizard 'AppSets' simultaneously; only one can be enabled. If you're opting for PixelSetupWizard, ensure that all packages under SetupWizard are also enabled, since PixelSetupWizard and SetupWizard share the same packages.

PR #10910, opened 2024-06-01
gasly-gapps

NihalShafi:
Hey there , Thanks for giving opportunity for us making our own gapps


nikgapps:

- Update 14/gasly.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #10985, opened 2024-06-11
Zed Gapps for OnePlus 5 and 5T

Zed9K:
Thanks a lot Nik <3


nikgapps:

- 14/Zed_OP5_5T doesn't have .config extension, we only accept config files!

PR #11008, opened 2024-06-16
coretts

fukiame:
just NikGapps Core with TTS, since there are users complaining about TTS being unavailable despite having it installed from Play Store


nikgapps:

- coretts.config must be part of Android Version folder, not outside of it!

PR #11039, opened 2024-06-18
Thanks Nik For your hardwork

black3093:


nikgapps:

- 14/Shubham Google Dailer Remover.config is not an aphanumeric name, make sure the name of config file is between A-Z and 0-9 additionally, accepted symbols are - (dash) or _ (underscore) any symbols including but not limited to (, ' . # ! *) or space are not accepted in the name try renaming the file to a name that doesn't contain any spaces or special characters
- Shubham Google Dailer Remover.config must be part of Android Version folder, not outside of it!

PR #11105, opened 2024-06-30
nifars-a14.config

rafin7667:
A14 gapps with minusing pixel thingy


nikgapps:

- config must be part of Android Version folder, not outside of it!

PR #11106, opened 2024-06-30
nifars_a14

rafin7667:
This gapps is created for Android 14 with skipping the pixel features and some few other things as well


nikgapps:

- Update 14/nifars-a14.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11141, opened 2024-07-04
sraajboy all gapps including addons

sraajboy:


nikgapps:

- Update 14/sraajboy_all.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 14/sraajboy_all.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11143, opened 2024-07-04
x3ll_gapps

x3ell:


nikgapps:

- Update 13/x3ll_ngapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11154, opened 2024-07-05
mygapps

kushagra055:


nikgapps:

- Update 14/mygapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #11195, opened 2024-07-12
dgapps.config

dcoxii1979:


nikhilmenghani:
@dcoxii1979 Meet is not officially released yet, please use GoogleDuo instead. It's running into error due to this.

PR #11223, opened 2024-07-16
sandeepandroid13.config

Mohangartoula:
This is create for android 13. For personal use.


nikgapps:

- Update 13/sandeepandroid13.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11243, opened 2024-07-17
A Android 11 NikGapps

ZXL-Q:


nikgapps:

- Update 11/nikgapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 11/nikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11265, opened 2024-07-19
begonia config

Faalcc:


nikgapps:

- Update 14/faalcc.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

nikhilmenghani:
@Faalcc you need to set the AppSet SetupWizard=0, you can keep the components under SetupWizard as 1 including >>SetupWizard

I will rename it in future releases to avoid confusion.

PR #11399, opened 2024-07-30
leonardogreen17.config

superprimeleo:
This Gapps is created for Android14. Please don't edit this gapps without my permission. Thank you for using my GAPPS.


nikgapps:

- leonardogreen17.config must be part of Android Version folder, not outside of it!

PR #11503, opened 2024-08-14
GappswithPL.config

sameer-dhiman:


nikgapps:

- Update 14/GappswithPL.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11642, opened 2024-08-27
nikgapps_pixelify_version 

Adnaanhashmi:
nikgapps which only contains gapps which gives pixelify experience with minimum gapps for AOSP roms.


nikgapps:

- Update 14/nikgapps_pixelify_version.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- nikgapps_pixelify_version.config must be part of Android Version folder, not outside of it!

PR #11674, opened 2024-08-31
essentials_aves

Aves210200:
replace gboard with heliboard after installing. Also, install mulch system webview


nikgapps:

- Update 14/core_with_slight_essentials.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #11888, opened 2024-10-02
hi

ngilani65:


nikgapps:

- Update 14/nima_moto_g_5g_plus.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- nima moto g 5g plus.config must be part of Android Version folder, not outside of it!

PR #11993, opened 2024-10-14
PS LESS SIZE NikGapps

Pawan-singh-PS:
Hey nik thanks again its interesting I am able to do this


nikhilmenghani:
Try not to add the comments in same line as the package definition.

Pixel launcher=1 # comment 

This will not work. Keep your comments in new line.

Also, you can't have spaces in the config name you're uploading. Please rename it to something small and refer other pull requests that are merged successfully.

PR #12150, opened 2024-11-07
GAppsMini-A10

gasleak08:
GAppsMini-A10


nikgapps:

- Update 10/GAppsGoMini-A10.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #12156, opened 2024-11-08
essential14

ameghna:


ameghna:
Hey Nikk. Thanks for your hard work.

nikgapps:

- 14/essential14.txt doesn't have .config extension, we only accept config files!
- Update 14/essential14.txt: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 14/essential14.txt, You cannot enable both Core and Core Go, you can only enable one of them

PR #12204, opened 2024-11-17
Create GosuGG

Gosu-Milktea:


nikgapps:

- 14/GosuGG doesn't have .config extension, we only accept config files!
- Update 14/GosuGG, You cannot enable both Core and Core Go, you can only enable one of them

PR #12239, opened 2024-11-23
DP-1897-gapps

DP1897:
Thanks for your hardwork


nikgapps:

- elite/15/DP-1897.config is not a valid path, make sure your elite config is in elite/<android_version>/<elite_folder_name>

PR #12241, opened 2024-11-23
DP-1897-gapps

DP1897:
Thanks for your hardwork


nikgapps:

- DP1897 is not allowed to upload to elite/15/DP-1897/DP-1897.config
- Update elite/15/DP-1897/DP-1897.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #12306, opened 2024-11-30
My-Personal-Gapps-A15

XxxEddiexxX:


nikgapps:

- 15/My-Personal-Gapps-A15 doesn't have .config extension, we only accept config files!
- Update 15/My-Personal-Gapps-A15: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #12363, opened 2024-12-05
nikgapps23

Muhammad-2311:
please check if it work


nikgapps:

- Update 14/nikgappslineage23.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #12438, opened 2024-12-14
Pixel Exp.config

sundarmanere:
This is Android 14 GAPPS which is built to provide a clean pixel experience.


nikgapps:

- Pixel Exp.config must be part of Android Version folder, not outside of it!

PR #12439, opened 2024-12-14
new a15 gapps

NihalShafi:


nikgapps:

- Cannot merge the changes automatically since status of 14/placeholder is removed, kindly start fresh with forking the repository again!
- 14/placeholder doesn't have .config extension, we only accept config files!
- Update 15/nihalgapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #12442, opened 2024-12-14
A14_gapps.config

sundarmanere:
This gapps are compatible with Android 14 providing pixel like experience in custom roms.


nikgapps:

- A14_gapps.config must be part of Android Version folder, not outside of it!

PR #12488, opened 2024-12-16
geojaz-gapps build

geojaz:


nikgapps:

- Update 12/geojaz-gapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 12/geojaz-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them
- Cannot merge the changes automatically since status of 12/placeholder is removed, kindly start fresh with forking the repository again!
- 12/placeholder doesn't have .config extension, we only accept config files!

PR #12567, opened 2024-12-25
Create a10.config

havi11:
I hope I don't mess up. Merry Christmas!


nikgapps:

- Update 14/a10.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

havi11:
Updated

PR #12606, opened 2024-12-30
A15-gapps-4me

chrysophilist:
Hi Nick,
Thanks for your hard work.


nikgapps:

- Update 15/A15Gapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #12755, opened 2025-01-12
nikgapps: q: some adjustment for nathanque.config

imnathanzero:
it was huge, I forgot to set to 0 for some packages


imnathanzero:
Also can you please remove the previous build ?

PR #12762, opened 2025-01-12
xyriins-build

krazy-xyriin:


nikgapps:

- 15/xyriins-build doesn't have .config extension, we only accept config files!

PR #12815, opened 2025-01-17
Fix

takeru-kageyuki:


nikgapps:

- 13/ProtonPlus_Fix doesn't have .config extension, we only accept config files!

PR #12909, opened 2025-01-27
nikgapps-by-vhonchon111

Vhonchon111:
hey nik, thanks for your hard work


nikgapps:

- Update 13/nikgapps-by-vhonchon111.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #13105, opened 2025-02-20
Spacewar---21022025

solomonrajan:
Spacewar---LineageOS---arm64---FridayBuild


nikgapps:

- Update 15/spacewar_20022025-lineageos.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #13133, opened 2025-02-23
nk47-pi4-gapps

Nuklear47:
Thanks for your hard work!
Without people like you alot of cool projects would not exist.


nikgapps:

- Update 15/nk47-pi4.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 15/nk47-pi4.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #13322, opened 2025-03-18
OnePlus7

sundarmanere:
[OnePlus7](https://github.com/sundarmanere/config/blob/main/15/OnePlus7)


nikgapps:

- Cannot merge the changes automatically since status of 15/OnePlus7 is renamed, kindly start fresh with forking the repository again!
- 15/OnePlus7 doesn't have .config extension, we only accept config files!

PR #13491, opened 2025-04-17
Add files via upload

trace-decay:
Enabled same packages as stock installer.


trace-decay:
Fixed. stock.config is within the 15 folder.

PR #13611, opened 2025-05-05
req

0xTaaa:


nikgapps:

- KONTOL.config must be part of Android Version folder, not outside of it!

PR #13615, opened 2025-05-05
nikgapps-rahul

Rahul867289:
hey nik thanks for the hard work.


nikgapps:

- Update 11/nikgapps-rahul.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #13636, opened 2025-05-08
new pull Request

roger010521:
Thanks for all the hardwork


nikgapps:

- Update 15/roger-gapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #13733, opened 2025-05-21
l20481141go gapps

leo20481141:
thank you very much nix team for allowing us to build custom gapps.


nikgapps:

- Update 14/l20481141go.config, You cannot enable both Core and Core Go, you can only enable one of them
- Update 15/l20481141go.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #13846, opened 2025-06-09
Add files via upload

kdrv3xc:


nikgapps:

- 15/msg4ws doesn't have .config extension, we only accept config files!
- 15/msg4ws is missing version information, please use latest version of NikGapps Config i.e. 37
- Update 15/pixelprimes_quindecim.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #14002, opened 2025-07-03
siam_android15.config

Siam-999:
This file is created for android 15 and most of the apps from OMNI pakage


Siam-999:
I will create another with correction

PR #14009, opened 2025-07-03
siam2_android15.config

mhs-mhs:
This file is created for android 15 devices and most of them from OMNI pakage.
Approve the one which don't have Gboard.


nikgapps:

- Update 15/siam2_android15.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #14031, opened 2025-07-07
cyph_a10.config

AIDEED09:
This GApps is created for Android 10 and has most of the apps from the Basic packages.


AIDEED09:
Hey Nik, thanks for your GApps!

PR #14054, opened 2025-07-12
Custom GApps Build

cypherab01:


nikgapps:

- .gitignore must be part of Android Version folder, not outside of it!
- Update 15/ab_nikgapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #14117, opened 2025-07-22
Add custom nikgapps.config with Play Store disabled

Boombastic747:
This config disables the Google Play Store package while keeping essential Google services enabled for Android 15.
It also resolves conflicts by enabling only one of SetupWizard or PixelSetupWizard (SetupWizard enabled here) and enabling only one Core package (Core enabled, CoreGo disabled) as per maintainer feedback.

Please review and consider merging this updated configuration. Thank you!

### DIFF
--- a/15/nikgapps.config
+++ b/15/nikgapps.config
@@ -1,0 +1,147 @@
+# NikGapps configuration file
+
+# If you are not sure about the config, just skip making changes to it or comment it by adding # before it
+# visit https://nikgapps.com/misc/2022/02/22/NikGapps-Config.html to read everything about nikgapps
+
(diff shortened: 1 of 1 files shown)


nikgapps:

- Update 15/nikgapps.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.
- Update 15/nikgapps.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #14197, opened 2025-08-11
Add a new Nikgapps for Android 15 on OnePlus Ace 5

ZXLQ114514:


nikgapps:

- Update 15/nikgappsgenko.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

PR #14233, opened 2025-08-17
coll147-gaaps

Coll147:


nikgapps:

- 15/lite-gaaps doesn't have .config extension, we only accept config files!

PR #14274, opened 2025-08-29
vanilla-p-F4

Professor-HASAN:
thanks


nikgapps:

- Update 16/vanilla-p-F4.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #14321, opened 2025-09-09
Create crdroid-poco.conf

ondrani:


nikgapps:

- 16/crdroid-poco.conf doesn't have .config extension, we only accept config files!

PR #14497, opened 2025-10-16
Minimal

RedLintu16:


RedLintu16:
Hey! Thank you for responding and checking my configuration file! I think it should be good to go now!

PR #14574, opened 2025-11-01
Add r7 config

kientrung0412:
Thanks for all


nikgapps:

- Update 14/r7-android-14.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #14693, opened 2025-11-27
nikgappsmeme

hudanazmul369-lgtm:
thanks


nikgapps:

- 16/nikgappsmeme.config.txt doesn't have .config extension, we only accept config files!
- The name 16/nikgappsmeme.config.txt is invalid. Please ensure the name consists only of letters (A-Z), numbers (0-9), dashes (-), or underscores (_). Special characters such as (, ' . # ! *), spaces, and others are not allowed. Try renaming the file to follow these guidelines.
- Update 16/nikgappsmeme.config.txt, You cannot enable both Core and Core Go, you can only enable one of them

PR #15283, opened 2026-04-22
Google Go Edition 

maxxcodebug:
Google Go Edition for android 16.


nikgapps:

- Update 16/AnxGoogleGoa16.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #15403, opened 2026-05-18
Bestgapps

mohdnajeeb333:


mohdnajeeb333:
Very Bad service... I requested already from more than 2 hour . But he didn't provide the gapps.

nikgapps:

- Update 11/Bestgappsmohdnajeeb333.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #15463, opened 2026-06-08
phpbb3 custom config

phpBB3:


nikgapps:

- Update 16/phpbb3.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

phpBB3:
@nikgapps This should be corrected now. If I understood right, the appset needs to be 0, but the packages under the appset still need to be 1 as they are needed by the addon PixelSetupWizard. Thank you.

PR #15489, opened 2026-06-16
NikGapp without Android Auto

sanpet:
NikGapp without Android Auto , no drive neither


nikgapps:

- Update 16/SSGApp-noAndroidAuto.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #15527, opened 2026-07-02
Added TodBonzalez config for 16

TodBonzalez:


nikgapps:

- elite/16/TodBonzalez.config is not a valid path, make sure your elite config is in elite/<android_version>/<elite_folder_name>

PR #15597, opened 2026-07-29
ShahabOS

XShahabDS:
hey nik thx for hardwork


nikgapps:

- Update 15/ShahabOS.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #15690, opened 2026-08-30
Updated config file

TodBonzalez:
If there is an option to keep SeedVault via config (e.g. `GoogleBackupTransport=2`) that would be ideal, but I'm not sure if the package has been updated yet so I left it out.

### DIFF
--- a/elite/16/TodBonzalez/TodBonzalez.config
+++ b/elite/16/TodBonzalez/TodBonzalez.config
@@ -54,9 +54,9 @@ Core=1
 >>GmsCore=1
 
 DigitalWellbeing=1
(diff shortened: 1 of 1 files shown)


nikgapps:

- Update elite/16/TodBonzalez/TodBonzalez.config: PixelSetupWizard and SetupWizard 'AppSets' cannot be enabled at the same time; only one can be active. If you choose PixelSetupWizard, make sure to enable all packages under SetupWizard as well, since PixelSetupWizard and SetupWizard use the same packages.

TodBonzalez:
Updated PixelSetupWizard to 0

PR #15696, opened 2026-08-31
ARTFULL

gptor26-Crip:


nikgapps:

- Update 16/ARTFULL.config, You cannot enable both Core and Core Go, you can only enable one of them

PR #15712, opened 2026-09-04
Added Google smaller components

igorkostic:


igorkostic:
Added uninstallable Google services

PR #15722, opened 2026-09-05
Lean build

igorkostic:
- No velvet
- Native weather and launcher


nikgapps:

- 16/rodin-crdroid-2launchers doesn't have .config extension, we only accept config files!
- 16/rodin-crdroid-gaaps-lean.conf doesn't have .config extension, we only accept config files!

PR #15741, opened 2026-09-07
Updated my version of crDroid gaaps

igorkostic:
Updated my version of crDroid gaaps


nikgapps:

- 16/igor-crdroid doesn't have .config extension, we only accept config files!

